PR #40, opened 2021-11-10
Fix relative oauth2 URLs

### LINKED ISSUE #39: Relative OAuth2 authorize/token URLs should resolve differently
According to the specs here: https://swagger.io/docs/specification/authentication/oauth2/#relative-url
```
servers:
  - url: https://api.example.com/v2
components:
  securitySchemes:
    oauth2sample:
      type: oauth2
      flows: 
        authorizationCode:
          authorizationUrl: /oauth/authorize  # <-----
          tokenUrl: /oauth/token              # <-----
          scopes:
            ...
```
should become
```
authorizationUrl: https://api.example.com/oauth/authorize
tokenUrl: https://api.example.com/oauth/token
```

But my experience is that `openapi-explorer` resolves them to
```
authorizationUrl: https://api.example.com/v2/oauth/authorize
tokenUrl: https://api.example.com/v2/oauth/token
```
I will open a PR shortly with a proposed change, hoping you'll consider it!

max-mathieu:
Fixes #39

### DIFF
--- a/src/templates/security-scheme-template.js
+++ b/src/templates/security-scheme-template.js
@@ -192,14 +192,15 @@ function oAuthFlowTemplate(flowName, clientId, clientSecret, apiKeyId, authFlow)
   let tokenUrl = authFlow.tokenUrl;
   let refreshUrl = authFlow.refreshUrl;
   const isUrlAbsolute = (url) => (url.indexOf('://') > 0 || url.indexOf('//') === 0);
+  const serverUrlObj = new URL(this.selectedServer.computedUrl);
   if (refreshUrl && !isUrlAbsolute(refreshUrl)) {
-    refreshUrl = `${this.selectedServer.computedUrl}/${refreshUrl.replace(/^\//, '')}`;
+    refreshUrl = `${serverUrlObj.origin}/${refreshUrl.replace(/^\//, '')}`;
   }
   if (tokenUrl && !isUrlAbsolute(tokenUrl)) {
-    tokenUrl = `${this.selectedServer.computedUrl}/${tokenUrl.replace(/^\//, '')}`;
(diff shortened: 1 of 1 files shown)


src/templates/security-scheme-template.js
```diff
@@ -192,14 +192,15 @@ function oAuthFlowTemplate(flowName, clientId, clientSecret, apiKeyId, authFlow)
   let tokenUrl = authFlow.tokenUrl;
   let refreshUrl = authFlow.refreshUrl;
   const isUrlAbsolute = (url) => (url.indexOf('://') > 0 || url.indexOf('//') === 0);
+  const serverUrlObj = new URL(this.selectedServer.computedUrl);
   if (refreshUrl && !isUrlAbsolute(refreshUrl)) {
-    refreshUrl = `${this.selectedServer.computedUrl}/${refreshUrl.replace(/^\//, '')}`;
```
wparad, line 196: I'm actually thinking for these three that we should follow the same pattern as the rest of the web and not custom logic here. The best way to do that would be:
```js
refreshUrl = new URL(refreshUrl, this.selectedServer.computedUrl);
```

And if we do this we can remove the method `isUrlAbsolute` because the `URL` takes care of that part too

wparad:
Thanks for this PR, I think adding this is a great idea. I have suggested improvement that we should make. And make sure you sign the CLA. - Thanks!

src/templates/security-scheme-template.js
max-mathieu: Nice, I didn't know about this constructor, let me update my PR!

max-mathieu:
Done. Also

(despite the link being wrong :) )

max-mathieu:
recheck

max-mathieu:

wparad:
Thank you!

PR #45, opened 2022-02-10
Add elapsed time

max-mathieu:
Display how long the API call took next to the response. This is useful for my team here as we're starting to look at our API performance, so maybe some other users could use it too

Some screenshots:
- a 200 
image
- a 500
image

It's not perfect (upload time is included), but getting detailed metrics on Fetch API is more work than I can afford right now

### DIFF
--- a/src/components/api-request.js
+++ b/src/components/api-request.js
@@ -704,7 +704,7 @@ export default class ApiRequest extends LitElement {
     const responseFormat = this.responseHeaders.includes('json') ? 'json' : (this.responseHeaders.includes('html') || this.responseHeaders.includes('xml')) ? 'html' : '';
     return html`
       <div class="row" style="font-size:var(--font-size-small); margin:5px 0">
-        <div class="response-message ${this.responseStatus}">Response Status: ${this.responseMessage}</div>
+        <div class="response-message ${this.responseStatus}">Response Status: ${this.responseMessage} in ${this.elapsedMs}ms</div>
         <div style="flex:1"></div>
         <button class="m-btn" part="btn btn-outline" @click="${this.clearResponseData}">CLEAR RESPONSE</button>
       </div>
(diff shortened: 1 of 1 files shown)


src/components/api-request.js
```diff
@@ -704,7 +704,7 @@ export default class ApiRequest extends LitElement {
     const responseFormat = this.responseHeaders.includes('json') ? 'json' : (this.responseHeaders.includes('html') || this.responseHeaders.includes('xml')) ? 'html' : '';
     return html`
       <div class="row" style="font-size:var(--font-size-small); margin:5px 0">
-        <div class="response-message ${this.responseStatus}">Response Status: ${this.responseMessage}</div>
+        <div class="response-message ${this.responseStatus}">Response Status: ${this.responseMessage} in ${this.elapsedMs}ms</div>
```
wparad, line 707: Just to make this a bit more clear I'm thinking about this small change
```suggestion
        <div class="response-message ${this.responseStatus}">Response Status: ${this.responseMessage}
           ${this.elapsedMs ? html`<span><br>Execution Time: ${this.elapsedMs}ms</span>` : ''}
         </div>
```

src/components/api-request.js
```diff
@@ -1050,7 +1050,9 @@ export default class ApiRequest extends LitElement {
       let respJson;
       let respText;
       tryBtnEl.disabled = true;
+      const fetchStart = new Date();
       fetchResponse = await fetch(fetchRequestObject);
+      this.elapsedMs = new Date() - fetchStart;
```
wparad, line 1055: To make sure this is reactive and shows up correctly, can we add the new property to:
* the constructor
* `clearResponseData`
* `get properties()` - (should be type of Number)

wparad (changes requested):
Overall I think this is a good idea.

I've left a few comments on how we can improve this to make it ready for release. Once those are places are updated, we can merge this in.

Thank you!

src/components/api-request.js
max-mathieu: Done

src/components/api-request.js
max-mathieu: Done -- hope I did this right :)

wparad (approved):

PR #53, opened 2022-05-23
There is a DoS vunlnerability in marked 2.0.1, so upgrading to 4.0.16

MoreNaruto:
Here's the vulnerability from Synk:

Screen Shot 2022-05-23 at 2 42 32 PM


MoreNaruto:
I have read the CLA Document and I hereby sign the CLA: Thomas Morris

wparad:
Thanks for finding this. I've reviewed it, and it works without any regressions.

Merging.

PR #88, opened 2022-09-02
multiple fixes

Alexeyca:


wparad:
Some of these look like great fixes, other ones are a bit hard to understand. From a tracking perspective, and getting your fixes merged in as soon as possible, we would appreciate if you could split this into multiple PRs each with an explicit expectation on what the purpose of the PR is.

One reason for doing this is so that we don't block the changes that should be made by the ones that still need to be discussed or have problems.

wparad (changes requested):
Please split these into multiple changes. You can use this PR to contain the first change as you see is most important.

PR #89, opened 2022-09-02
fixing Request Body button in parameters.

Alexeyca:


src/components/api-request.js
```diff
@@ -461,7 +461,7 @@ export default class ApiRequest extends LitElement {
                 <button class="tab-btn ${this.activeSchemaTab === 'body' ? 'active' : ''}" data-tab="body">BODY</button>
               </div>
               ${html`<div class="tab-content col" style="display: ${this.activeSchemaTab === 'model' ? 'block' : 'none'}"> ${reqBodySchemaHtml}</div>`}
-              ${html`<div class="tab-content col" style="display: ${this.activeSchemaTab === 'model' ? 'none' : 'block'}"> ${reqBodyExampleHtml}</div>`}
+              ${html`<div class="tab-content col" style="display: ${this.activeSchemaTab === 'body' ? 'none' : 'block'}"> ${reqBodyExampleHtml}</div>`}
```
wparad, line 464: This change doesn't seem right, what's the problem that this is fixing?

src/components/api-request.js
```diff
@@ -539,9 +539,9 @@ export default class ApiRequest extends LitElement {
               : html`
                 ${fieldType === 'object'
                   ? html`
-                  <div class="tab-panel row" style="min-height:220px; border-left: 6px solid var(--light-border-color); align-items: stretch;">
+                  <div class="tab-panel row" style="min-height:300px; border-left: 6px solid var(--light-border-color); align-items: stretch;">
```
wparad, line 542: Can you share a bit about what's the reason for this change, maybe a couple of pictures or an example spec where we can trade off on what this size should actually be. Obviously it's hard coded for a reason so figuring out the right value for this is always a challenge.

wparad (commented):
I have two questions about these changes, hopefully they make sense.

src/components/api-request.js
Alexeyca: Did you try to use it? it not enough room for Request Body button, and that button move under schema

wparad: Do you have a sample spec I can use to validate the change?

Alexeyca: Screen Shot 2022-09-02 at 9 36 31 AM

Alexeyca: technically, it totally not usable

wparad: Next time I would appreciate a sample spec to validate this. From what I've been able to test it seems to okay.

wparad (approved):
I've approved this, but it won't be able to be merged until the CLA is signed.

Alexeyca:

Alexeyca:

PR #90, opened 2022-09-02
fixing array types in parameters.

Alexeyca:


Alexeyca:
Screen Shot 2022-09-02 at 12 54 49 PM

Alexeyca:

src/components/api-request.js
```diff
@@ -500,7 +500,9 @@ export default class ApiRequest extends LitElement {
             <div class="param-name ${fieldSchema.deprecated ? 'deprecated' : ''}">
               ${fieldName}${!fieldSchema.deprecated && (schema.required && schema.required.includes(fieldName) || fieldSchema.required) ? html`<span style='color:var(--red);'>*</span>` : ''}
             </div>
-            <div class="param-type">${paramSchema.type}</div>
+            <div class="param-type">
+            ${paramSchema.type === 'array' ? html`[<span class="formnum">${paramSchema.format}</span>]` : `${paramSchema.type}`}
```
wparad, line 504: If it is an array of booleans, I don't see this working correctly, since you are making a similar change in the schema-tree, I would normally expect a similar solution here.

src/components/schema-tree.js
```diff
@@ -244,7 +244,12 @@ export default class SchemaTree extends LitElement {
                 : ''
           }
           <span class="${dataTypeCss}">
-            ${dataType === 'array' ? `${type}[]` : `${type}`}
+            ${dataType === 'array'
```
wparad, line 247: Let's not change the format from `type[]` to `[type]`, so let's revert that change.

Alexeyca:
This fix about to represent array more informative. Because explanation as ARRAY is provide you no idea array of what?

src/components/schema-tree.js
```diff
@@ -244,7 +244,12 @@ export default class SchemaTree extends LitElement {
...
+            ? type.match(/integer|number/g)
+            ? html`[<span class="formnum">${format}</span>]`
+            : `[${type}]` : type.match(/integer|number/g)
+            ? html`${type}(<span class="formnum">${format}</span>)`
```
wparad, line 251: There is already a number class, let's reuse that one instead `number` or `integer` in the `class`, not hard coded to a new class.

src/utils/schema-utils.js
```diff
@@ -25,6 +25,10 @@ export function getTypeInfo(schema) {
     dataType = IS_MISSING_TYPE_INFO_TYPE;
   }
 
+  if (dataType === 'array' && !schema.format) {
+    schema.format = schema.items.format;
```
wparad, line 29: let's not change the `schema.format` but instead populate `info.arrayItem.format` like we do on line 51 so that the code is consistent.

src/components/schema-tree.js
```diff
@@ -224,7 +224,7 @@ export default class SchemaTree extends LitElement {
     }
 
     // For Primitive Data types
-    const { type, readOrWriteOnly, constraint, defaultValue, example, allowedValues, pattern, schemaDescription, schemaTitle, deprecated } = JSON.parse(data);
+    const { type, format, readOrWriteOnly, constraint, defaultValue, example, allowedValues, pattern, schemaDescription, schemaTitle, deprecated } = JSON.parse(data);
```
wparad, line 227: Any change we make in this file needs to also be made in the `schema-table.js` component as well.

src/styles/schema-styles.js
```diff
@@ -58,6 +58,7 @@ export default css`
 
 .stri, .string, .uri, .url, .byte, .bina, .date, .pass, .ipv4, .ipv4, .uuid, .emai, .host {color:var(--green);}
 .inte, .numb, .number, .int6, .int3, .floa, .doub, .deci .blue {color:var(--blue);}
+.formnum {color:var(--orange);}
```
wparad, line 61: orange is for booleans.

src/utils/schema-utils.js
```diff
@@ -76,6 +80,7 @@ export function getTypeInfo(schema) {
   info.constraint = constraint;
   info.html = JSON.stringify({
     type: info.type,
+    format: info.format,
```
wparad, line 83: Do we need this in the `html` property if we use the `info.arrayItemFormat` property?

wparad (changes requested):
There's a lot here still, I tried to comment on each change, let me know if anything doesn't make sense.

src/components/api-request.js
```diff
@@ -580,9 +582,9 @@ export default class ApiRequest extends LitElement {
                           data-pname = "${fieldName}"
                           data-default = "${paramSchema.default || ''}"
                           spellcheck = "false"
-                        >${this.fillRequestWithDefault === 'true' ? paramSchema.default : ''}</textarea>
+                        >${this.fillRequestWithDefault === 'true' ? formdataPartExample[0] && formdataPartExample[0].exampleValue : ''}</textarea>
```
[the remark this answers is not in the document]
Alexeyca: do you mean if i have a parameter with 200 lines, I must somehow remember and manually type or copy/paste it from somewhere?
Just to make it clear, i not do it for fun. We have a project where we want to use it, and this is what our team requiring to better usage.

src/components/schema-tree.js
Alexeyca: why? how i will notice if this is array?

src/components/schema-tree.js
Alexeyca: i need different colour here to notice exact type of integer

src/utils/schema-utils.js
Alexeyca: yep, we need format of Integer, format of Array

src/components/api-request.js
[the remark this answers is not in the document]
wparad: Let's revert this in this PR, as it currently is, this is for the default property not for the example. If there is a concrete problem, can you open an issue in this repo so we can discuss the specific problem, that way we can be aligned before jumping into a solution.

src/components/schema-tree.js
wparad: What do you mean? `type` is a type and `type[]` is an array.

src/components/schema-tree.js
wparad: This is potentially going to be hard sell for others. Can we focus on fixing the display of the array and type in this PR and level coloring it differently to a later PR?

src/components/schema-tree.js
Alexeyca: if i correct, this part will be similar to swagger-ui design, and it will more readable

src/styles/schema-styles.js
Alexeyca: Open to any suggestion. What colour do you prefer to notice format?

wparad: Let's not change the color in this PR, we can discuss in a follow up once we fix the array formatting.

src/components/api-request.js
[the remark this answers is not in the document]
Alexeyca: Do you want an example request?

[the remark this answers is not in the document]
wparad: In a github issue, yes please

src/utils/schema-utils.js
Alexeyca: Screen Shot 2022-09-02 at 1 49 02 PM

src/components/api-request.js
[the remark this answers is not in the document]
Alexeyca: i have no time to create an issue, sorry.
You can put any large JSON in parameter and try to type it manually later.  And imagine how you will handle this.
But this is optional! So you can disable it.

src/styles/schema-styles.js
Alexeyca: Screen Shot 2022-09-02 at 1 49 02 PM

src/components/schema-tree.js
wparad: Let's not make visual changes just because "we like swagger-ui" better. If we want to change the display in a later PR we can do that, let's fix the array/format display here.

src/components/api-request.js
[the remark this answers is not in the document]
wparad: I'm not going to argue this point in this PR. Please decide which feature you would like the most and stick to that in this PR. 

Here is your issue for you to make your case why this should be changed: https://github.com/Rhosys/openapi-explorer/issues/91

Until then, please remove this code from the PR.

src/styles/schema-styles.js
wparad: What are you trying to show me here?

src/components/schema-tree.js
Alexeyca: if we did like swagger-ui we use swagger-ui. But we try to finish one of fork that will be useful from point of documentation usability. If you think we are in the wrong fork, that ok. You are the maintainer, this is your rights.

src/styles/schema-styles.js
Alexeyca: nothing more. 
Look like i wasting time in useless discussions that most not about code quality, but about how to keep tool not informative, because some good solution will remind about Swagger. :)
You can take "Swagger Petstore - OpenAPI 3.0" and test your tool. 
Yes, this is from Swagger, but to check API that built not by you will give you pile of experience about info that your tool provide and what not.
Let me know if i can somehow help

wparad: It's about context. I appreciate the PR the problem is that we aren't on the same page yet. You have things that you see, and you are trying to fix them but without sharing the context. It would really help instead of a PR to start with the issue and then we can go from there:
* Here is problem X
* Here is problem Y
Of which we can talk about each one in the PR and figure out the best solution for everyone.

Right now we are talking about the solution before agreeing about the problem.

src/components/schema-tree.js
wparad: It's a problem of understanding. It sounds like you are saying "We should change this because Swagger-ui does it this way", which doesn't help me understand the problem. It's possible there is an issue with our current solution, just help me understand what that is.

It seems like you are suggesting `[type]` is more readable than `type[]`. Perhaps it is, I guess from a naive perspective one could say, I don't see the difference. Which makes it subjective. If this change is what sells using this library for you, help us understand how to justify one display over the other. We want to pick the best one, objectively.

Alexeyca: For me, it looks like you self never use your tool. Can you notice a difference?
Screen Shot 2022-09-02 at 3 01 51 PM
Screen Shot 2022-09-02 at 3 04 01 PM

Alexeyca: and simple way to check changes, just switch to branch, build and check in browser

wparad: I had a talk with our UX team, let's go the `[int64]`, but the `[` and `]` should be white/default color, and then the `int64` color can be left to be blue as defined by the `integer` class.

Make sense?

src/components/schema-tree.js
wparad: I like your idea of override `type` with the format, so we shouldn't need to pass this property in. Does that make sense?

wparad:
Deploy these changes in release: `0.9.381`

PR #115, opened 2022-12-07
Add support for "pattern properties" in schema display

IMSoP:
JSON Schema (and therefore OpenAPI 3.1) includes support for "pattern properties", which match properties based on a regex. This patch adds these to the schema view in a similar style to "additional properties", e.g. pattern '^abc-\d+$' shows as '<pattern: ^abc-\d+$>'

https://json-schema.org/understanding-json-schema/reference/object.html#pattern-properties


IMSoP:

wparad:
`0.10.446`

PR #122, opened 2023-01-13
Fix #121 Improve enum support in response example generation.

### LINKED ISSUE #121: Generated example responses do not account for non string enums
I didn't dig very deep into the code but from a quick scan of `getSampleValueByType` in `schema-utils.js`, it looks like there is no logic to account for enums other than string. As of 307e0a8.
```js
if (typeValue.match(/^string/g)) {
    if (schemaObj.enum) { return schemaObj.enum[0]; }
    // ...
}
```

Example response model will output;
```
number                               Allowed: 404┃500
```
But the example will output;
```
0
```

Minimum reproducible example;
```json
{
  "openapi": "3.1.0",
  "paths": {
    "/example": {
      "get": {
        "responses": {
          "200": {
            "description": "Example response",
            "content": {
              "plain/text": {
                "schema": {
                  "type": "number",
                  "enum": [
                    404,
                    500
                  ]
                }
              }
            }
          }
        }
      }
    }
  }
}
```

darcyrush:
Feel free to modify; I haven't tested extensively so I have probably missed some edge cases. Changes are to generate more accurate examples for types other than string.

Example response Model;
```
{
    a: string
    b: boolean
    c: number enum            Allowed: 111┃222
    d: email
    e: boolean enum           Allowed: true
    f: date
    g: number
    h: string enum            Allowed: some┃string
}
```

Example response Example;
```
{
    a: "a"
    b: false
    c: 111
    d: "user@example.com"
    e: true
    f: "2023-01-13"
    g: 0
    h: "some"
}
```
Test schema;
```json
{
  "openapi": "3.1.0",
  "paths": {
    "/example": {
      "get": {
        "responses": {
          "200": {
            "description": "Example response",
            "content": {
              "application/json": {
                "schema": {
                  "type": "obect",
                  "properties": {
                    "a": {
                      "type": "string",
                    },
                    "b": {
                      "type": "boolean",
                    },
                    "c": {
                      "type": "string",
                      "enum": [
                        111,
                        222
                      ]
                    },
                    "d": {
                      "type": "string"
                      "format": "email"
                    },
                    "e": {
                      "type": "boolean",
                      "enum": [
                        true
                      ]
                    },
                    "f": {
                      "type": "string",
                      "format": "date"
                    },
                    "g": {
                      "type": "number",
                    },
                    "h": {
                      "type": "string",
                      "type": "string",
                      "enum": [
                        "some",
                        "string"
                      ]
                    },
                  }
                }
              }
            }
          }
        }
      }
    }
  }
}
```

### DIFF
--- a/src/styles/schema-styles.js
+++ b/src/styles/schema-styles.js
@@ -56,7 +56,7 @@ export default css`
 .inte, .integer, .numb, .number, .int6, .int64, .int3, .int32, .floa, .float, .doub, .double, .deci, .decimal, .blue { color:var(--blue); }
 .null { color:var(--red); }
 .bool, .boolean { color:var(--orange); }
-[class$="enum"], .cons, .const { color:var(--yellow); }
+.enum, .cons, .const { color:var(--yellow); }
 
 .tree .toolbar {
   display: flex;
--- a/src/styles/schema-styles.js
+++ b/src/styles/schema-styles.js
@@ -56,7 +56,7 @@ export default css`
 .inte, .integer, .numb, .number, .int6, .int64, .int3, .int32, .floa, .float, .doub, .double, .deci, .decimal, .blue { color:var(--blue); }
 .null { color:var(--red); }
 .bool, .boolean { color:var(--orange); }
-.enum, .cons, .const { color:var(--yellow); }
+[class$="enum"], .cons, .const { color:var(--yellow); }
 
 .tree .toolbar {
   display: flex;
--- a/src/utils/schema-utils.js
+++ b/src/utils/schema-utils.js
@@ -19,7 +19,9 @@ export function getTypeInfo(schema, options = { includeNulls: false }) {
   } else if (schema.type) {
     const arraySchema = Array.isArray(schema.type) ? schema.type : (typeof schema.type === 'string' ? schema.type.split('┃') : schema.type);
     dataType = Array.isArray(arraySchema) ? arraySchema.filter((s) => s !== 'null' || options.includeNulls).join('┃') : schema.type;
-    dataType = dataType.replace('string', typeof schema.const !== 'undefined' && 'const' || schema.enum && 'enum' || schema.format || 'string');
+    ['string', 'number'].forEach(type => {
+      dataType = dataType.replace(type, typeof schema.const !== 'undefined' && 'const' || schema.enum && `${type} enum` || schema.format || type);
+    });
 
     if (schema.nullable && options.includeNulls) {
       dataType += '┃null';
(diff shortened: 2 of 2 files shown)


darcyrush:

darcyrush:
recheck

src/utils/schema-utils.js
```diff
@@ -19,7 +19,9 @@ export function getTypeInfo(schema, options = { includeNulls: false }) {
   } else if (schema.type) {
     const arraySchema = Array.isArray(schema.type) ? schema.type : (typeof schema.type === 'string' ? schema.type.split('┃') : schema.type);
     dataType = Array.isArray(arraySchema) ? arraySchema.filter((s) => s !== 'null' || options.includeNulls).join('┃') : schema.type;
-    dataType = dataType.replace('string', typeof schema.const !== 'undefined' && 'const' || schema.enum && 'enum' || schema.format || 'string');
+    ['string', 'number', 'boolean'].forEach(type => {
+      dataType = dataType.replace(type, typeof schema.const !== 'undefined' && 'const' || schema.enum && `${type} enum` || schema.format || type);
+    })
```
wparad, line 24:
```suggestion
    });
```

src/utils/schema-utils.js
```diff
@@ -19,7 +19,9 @@ export function getTypeInfo(schema, options = { includeNulls: false }) {
   } else if (schema.type) {
     const arraySchema = Array.isArray(schema.type) ? schema.type : (typeof schema.type === 'string' ? schema.type.split('┃') : schema.type);
     dataType = Array.isArray(arraySchema) ? arraySchema.filter((s) => s !== 'null' || options.includeNulls).join('┃') : schema.type;
-    dataType = dataType.replace('string', typeof schema.const !== 'undefined' && 'const' || schema.enum && 'enum' || schema.format || 'string');
+    ['string', 'number', 'boolean'].forEach(type => {
```
wparad, line 22: According to the spec, it has to be a `string`: https://spec.openapis.org/oas/latest.html#serverVariableEnum I'm okay with adding number support here, but it doesn't make sense for boolean, so let's pull that one out.

```suggestion
    ['string', 'number'].forEach(type => {
```

src/styles/schema-styles.js
```diff
@@ -56,7 +56,7 @@ export default css`
 .inte, .integer, .numb, .number, .int6, .int64, .int3, .int32, .floa, .float, .doub, .double, .deci, .decimal, .blue { color:var(--blue); }
 .null { color:var(--red); }
 .bool, .boolean { color:var(--orange); }
-.enum, .cons, .const { color:var(--yellow); }
+[class$="enum"], .cons, .const { color:var(--yellow); }
```
wparad, line 59: There's a bug with the enum css type, we should fix it there instead of here, I'll add a comment with the suggested fix.

wparad (changes requested):
This is really interesting, while it is against the spec to have this as far as I can see, I don't think it is a bad addition. So let's do it. I've made some small comments, and once they're resolved we should be good to go.

Re: the CLA bot, I believe the bot is complaining that the email on the commit `drush@cbhbank.com` isn't part of your user profile. Haven't seen this problem before. If you don't want to update your profile (or change the commit to be a valid email) that isn't a problem, we can still merge the PR, once the fixes are in.

darcyrush:
Huh. I didn't realise it wasn't in the spec since I remembered some level of support for it in other generators/viewers. I don't mind if you close this without merging in that case.

Requested changes have been pushed anyway.

src/styles/schema-styles.js
wparad:
```suggestion
.enum, .cons, .const { color:var(--yellow); }
```

wparad (approved):

wparad:
Available in `0.11.481`

PR #125, opened 2023-02-10
Display descriptions for URL parameters

IMSoP:
"Parameter Objects" in the OpenAPI spec can include a "description" property, which may include CommonMark formatting: https://spec.openapis.org/oas/latest.html#parameter-object

### DIFF
--- a/src/components/api-request.js
+++ b/src/components/api-request.js
@@ -210,6 +210,13 @@ export default class ApiRequest extends LitElement {
         ${this.renderStyle === 'focused'
           ? html`
             <td>
+              ${param.description
+                ? html`
+                  <div class="param-description">
+                      ${unsafeHTML(marked(param.description))}
+                  </div>`
+                : ''
+              }
               ${paramSchema.default || paramSchema.constraint || paramSchema.allowedValues || paramSchema.pattern
                 ? html`
                   <div class="param-constraint">
--- a/src/styles/api-request-styles.js
+++ b/src/styles/api-request-styles.js
@@ -32,6 +32,17 @@ export default css`
 .api-request .param-constraint:empty {
   display:none;
 }
+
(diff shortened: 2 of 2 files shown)


wparad:
Interesting, it seems we weren't displaying this, so I appreciate the PR.

I'm looking at the update, and I'm trying to wrap my head around if there is something we can do to improve the UI here. I see that the text in the right cell is off because the new description has `<p>` element:
```html
![image](https://user-images.githubusercontent.com/5056218/218171686-0c7cb705-7dac-4276-bd80-33ae0457c488.png)

File: src/components/api-request.js [hunk 1]
```diff
@@ -238,6 +238,13 @@ export default class ApiRequest extends LitElement {
                   </div>`
                 : ''
               }
+              ${param.description
```
wparad, line 241: I'm thinking we should swap this up above the `param-constraint` block.

src/styles/api-request-styles.js
```diff
@@ -32,6 +32,14 @@ export default css`
 .api-request .param-constraint:empty {
   display:none;
 }
+
+.api-request .param-description {
+  min-width:100px;
+}
```
wparad, line 38:
```suggestion
}
.api-request .param-description p {
  margin-block: 0 0.5em;
}
```

wparad (changes requested):
Two small things and then we should good to merge. If we swap the block order and apply the margin based on my comment, I think this will look much better:

image

IMSoP:
Yes, I wasn't sure whether it should go above or below the constraints, but testing more, I agree it looks better in the other order. I've applied that and the extra margin rule.

wparad (approved):

wparad:
That has been released in version `0.11.487`

PR #127, opened 2023-02-10
Don't use monospace font for property descriptions

IMSoP:
Currently, object and array property descriptions are in regular font, but others are forced to monospace with an inline style.

In my opinion, the monospace font looks ugly here, since descriptions are generally human-readable text, not code, so this PR removes the inline style.

However, if this is a desirable feature, let me know, and I will work on an alternative approach where all descriptions are styled consistently, and can be over-ridden in some way by the user.

### DIFF
--- a/src/components/schema-table.js
+++ b/src/components/schema-table.js
@@ -238,7 +238,7 @@ export default class SchemaTable extends LitElement {
         </div>
         <div class='td key-descr'>
           ${dataType === 'array' ? html`<span class="m-markdown-small">${unsafeHTML(marked(description))}</span>` : ''}
-          <span class="m-markdown-small" style="font-family: var(--font-mono); vertical-align: middle;">
+          <span class="m-markdown-small" style="vertical-align: middle;">
             ${unsafeHTML(marked(`${dataType === 'array' && description || `${schemaTitle ? `**${schemaTitle}:**` : ''} ${schemaDescription}` || ''}`))}
           </span>
           ${this.schemaDescriptionExpanded ? html`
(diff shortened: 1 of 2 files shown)


wparad:
Do you have a before and after screenshot of what is being changed. It sounds like a good idea from your description, I just want to double check with visual validation.

IMSoP:
Sure; I'll try to include examples in future PRs. The below are all from this sample schema: https://gist.github.com/IMSoP/693a511bf205d282c59c0929f1fe4222

You also need to set `--font-mono` for the difference to be visible; the actual fonts I'm using are from Google Fonts:

```html
    <link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,300;0,700;1,300;1,700&display=swap">
    <link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=Roboto+Mono:wght@300;700&display=swap">
    <style>
        openapi-explorer {
            --font-regular: Roboto, "Open Sans", sans-serif;
            --font-mono: 'Roboto Mono', monaco, monospace;
        }
    </style>
```

Here's what it currently looks like:

1--schema-table-mixed-font

This is what I'm proposing:

2--schema-table-proposed-variable-font

Alternatively, we could remove the inline styles but make monospace consistent, but I think the result is rather ugly:

3---schema-table-alternative-fixed-font

I'm less sure about which font makes most sense in the "tree" view; we'll need to decide before merging, because it currently shares CSS.

Monospace tree:

4--schema-tree-fixed-font

Variable font tree:

5--schema-tree-variable-font

wparad (approved):
With the screenshots it's an obvious improvement, I'm approving and merging this in.

PR #143, opened 2023-03-14
Use animation rather than transition to avoid arbritrary max-height

IMSoP:
For requests and repsonses with a lot of properties, the schema and examples are currently truncated, because the containing elements have max-height set to 5000px or 10000px.

This value is part of the expand/collapse transition - animating from zero to 'auto' turns out to be a hard problem, as discussed in this Stack Overflow question: https://stackoverflow.com/q/3508605/157957

This patch takes an alternative approach: using "animation" rather than "transition", the last frame of the animation can have a max-height, but then revert to auto. The "snap" to full height will happen off-screen, unless your screen is more than 5000px tall.

This also allows a slightly more natural collapse transition, quickly collapsing long blocks, so that shorter blocks don't have an awkward pause before they start animating.

### DIFF
--- a/src/components/json-tree.js
+++ b/src/components/json-tree.js
@@ -55,13 +55,30 @@ export default class JsonTree extends LitElement {
         border: 1px solid var(--border-color);
       }
 
+      /* Expand a block from 0 to something visible, then to probably full screen, then beyond most screens.
+         Unlike with a transition, the final CSS can have an unbounded height, which will take effect after the animation. */
+      @keyframes expand-height {
+        0% { max-height: 0; }
+        50% { max-height: 100px; }
+        95% { max-height: 1000px; }
+        100% { max-height: 5000px; }
+      }
+      /* Inverse of the above, collapsing quickly at first (to avoid a delay if the element is already quite short)
+         then slowing towards 0. */
+      @keyframes collapse-height {
+        0% { max-height: 5000px; }
+        5% { max-height: 500px; }
+        50% { max-height: 100px; }
+        100% { max-height: 0; }
+      }
       .inside-bracket-wrapper {
-        max-height: 10000px;
-        transition: max-height 1.2s ease-in-out;
         overflow: hidden;
       }
+      .open-bracket:not(.collapsed) + .inside-bracket-wrapper {
+        animation: linear 0.2s expand-height;
+      }
       .open-bracket.collapsed + .inside-bracket-wrapper {
-        transition: max-height 1.2s ease-in-out -1.1s;
+        animation: linear 0.2s collapse-height;
         max-height: 0;
       }
       .inside-bracket {
--- a/src/components/json-tree.js
+++ b/src/components/json-tree.js
@@ -4,6 +4,7 @@ import { getI18nText } from '../languages';
 import FontStyles from '../styles/font-styles.js';
 import BorderStyles from '../styles/border-styles';
 import InputStyles from '../styles/input-styles';
+import KeyFrameStyles from '../styles/key-frame-styles.js';
 
 export default class JsonTree extends LitElement {
   static get properties() {
@@ -18,6 +19,7 @@ export default class JsonTree extends LitElement {
       FontStyles,
       BorderStyles,
       InputStyles,
+      KeyFrameStyles,
       css`
       :host{
         display:flex;
@@ -55,22 +57,6 @@ export default class JsonTree extends LitElement {
         border: 1px solid var(--border-color);
       }
 
-      /* Expand a block from 0 to something visible, then to probably full screen, then beyond most screens.
-         Unlike with a transition, the final CSS can have an unbounded height, which will take effect after the animation. */
-      @keyframes expand-height {
-        0% { max-height: 0; }
-        50% { max-height: 100px; }
-        95% { max-height: 1000px; }
-        100% { max-height: 5000px; }
-      }
-      /* Inverse of the above, collapsing quickly at first (to avoid a delay if the element is already quite short)
-         then slowing towards 0. */
-      @keyframes collapse-height {
-        0% { max-height: 5000px; }
-        5% { max-height: 500px; }
-        50% { max-height: 100px; }
-        100% { max-height: 0; }
-      }
       .inside-bracket-wrapper {
         overflow: hidden;
       }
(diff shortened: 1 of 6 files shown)


src/components/json-tree.js
```diff
@@ -55,13 +55,30 @@ export default class JsonTree extends LitElement {
...
+         Unlike with a transition, the final CSS can have an unbounded height, which will take effect after the animation. */
+      @keyframes expand-height {
+        0% { max-height: 0; }
+        50% { max-height: 100px; }
+        95% { max-height: 1000px; }
+        100% { max-height: 5000px; }
+      }
+      /* Inverse of the above, collapsing quickly at first (to avoid a delay if the element is already quite short)
+         then slowing towards 0. */
+      @keyframes collapse-height {
```
wparad, line 68: I feel like there was a "reverse" for key frames somewhere, does that not actually exist?

src/styles/schema-styles.js
```diff
@@ -95,6 +95,22 @@ export default css`
 }
 .toolbar-item:first-of-type { margin:0 2px 0 0;}
 
+/* Expand a block from 0 to something visible, then to probably full screen, then beyond most screens.
```
wparad, line 98: Any thoughts of where we can put this to make it reusable, I don't have a problem duplicating it, but I think we actually want all of these to feel the same, right?

IMSoP: Yeah, I managed to reduce to two copies rather than three, but couldn't find anywhere that was shared between all three contexts. That may just be because I don't know the structure of the repo, though.

src/components/json-tree.js
IMSoP: There is an `animation-direction` property ... but it doesn't work for this use case. The problem is that if you switch between two classes, and the only difference is `animation-direction`, the animation doesn't restart - the browser just says "well, I've already played the animation for this element, so direction is irrelevant".

If the two classes have different `animation-name` properties (referencing different key frames), then the browser says "ooh, a new animation has been assigned to this element, I'll start it".

(This took me some time to discover and get my head around.)

src/styles/schema-styles.js
wparad: I honestly don't know that there is, maybe it's worth just creating that file, you can find the pattern from the style related imports into these classes

```css
static finalizeStyles() {
    return [
      FontStyles,
      SchemaStyles,
      BorderStyles,
      ...
```

So you could import another one, like `keyframeStyles` or whatever. It's not a big deal, but I think it's worth it.

src/components/json-tree.js
wparad: Hmm weird. I wonder if the `<details>` html tag solves this problem, haven't really looked into that one, but that triggered that though. No matter, I can understand that making sense.

wparad: Hmm weird. I wonder if the `<details>` html tag solves this problem, haven't really looked into that one, but that triggered that though. No matter, I can understand that making sense.

IMSoP:
> Interesting, do you have a spec that I can see this in action? I'd be curious see the impact on just arbitrarily setting the max-height to be arbitrarily larger.

Well, that was my first thought; but as you say under #126, that's just putting off the problem, because there's no natural limit on how long a fully-expanded spec or example "should" be.

Also, the larger you make the max-height, the harder it is to make the transition look smooth for smaller elements - most of the transition ends up spent changing `max-height` between a few hundred pixels and many thousands, with no visible effect on the element.

src/styles/schema-styles.js
IMSoP: Ah, yes, I'll give that a try.

IMSoP: Yep, that seems to work; I've moved the existing "spin" animation in there as well, just seems logical.

wparad: Awesome.

src/styles/key-frame-styles.js
```diff
@@ -0,0 +1,25 @@
...
+     Unlike with a transition, the final CSS can have an unbounded height, which will take effect after the animation. */
+  @keyframes expand-height {
+    0% { max-height: 0; }
+    50% { max-height: 100px; }
+    95% { max-height: 1000px; }
+    100% { max-height: 5000px; }
+  }
+  /* Inverse of the above, collapsing quickly at first (to avoid a delay if the element is already quite short)
+     then slowing towards 0. */
+  @keyframes collapse-height {
```
wparad, line 19: I think there is something off with the collapse, it waits until the animation of the rotating icon is done before collapsing, which makes the animation feel stuck. Any idea on what the difference is here?

waits-until-animation-is-done

wparad (commented):
I found one thing with the collapsing which feels weird, hoping that's the last thing and then we can merge this in.

src/styles/key-frame-styles.js
IMSoP: It's not technically waiting for anything, it's just that all the values of `max-height` from 5000px down to 100px or so don't make any visible difference, so you have to wait for all those animation frames before the ones that actually collapse it.

That's presumably why the existing `transition` has a negative delay (the `-1.0s`) which skips over that part of the transition. The `keyframes` tries to do something similar with the extra stops at 5% and 50%, so we can tweak those to be more aggressive if it still feels laggy - maybe `100px` should be more like 25%? The only tradeoff is making sure larger elements still look smooth.

wparad: What do you think of this?

```css
@keyframes collapse-height {
    0% { max-height: 5000px; }
    5% { max-height: 500px; }
    40% { max-height: 100px; }
    85% { max-height: 0; }
  }
```

IMSoP: Setting to 0 before reaching 100% feels weird - you can just make the absolute duration on the `animation` property shorter in that case.

Maybe we just need an extra key frame between 100px and 0? I've just shut my laptop down for the night so can't test, but halfway between your last two frames would be 50px at 62.5%, so maybe something like this?

```
@keyframes collapse-height {
    0% { max-height: 5000px; }
    5% { max-height: 500px; }
    40% { max-height: 100px; }
    60% { max-height: 50px; }
    100% { max-height: 0; }
}
```

IMSoP: I thought of two ways to improve the feel of this: https://github.com/Rhosys/openapi-explorer/pull/143/commits/e5dbb7515124e92cb0340ac3b0b3b8221c399559

Have a play and see if that feels a bit more reactive.

wparad: I think this is great, let's go with it.

PR #145, opened 2023-04-05
Use default fg and bg2 CSS variables in json-tree and schema-tree

### LINKED ISSUE #144: Add CSS variables for custom text and background colors for schema-tree and json-tree
It would be nice to allow custom colors for the schema-tree and json-tree components. Maybe introduce variables just like for the navbar.

In schema-tree.js and json-tree.js replace:

```
background: var(--primary-color);
color: white;
```

with:

```
background: var(--tree-bg-color);
color: var(--tree-text-color);
```

When background is primary color this is how it currently look like:
before

This is how I want it to look like with a custom text and background color:
after

bergander:
Fixed #144. By removing the color property it will inherit --fg from host instead. Also remove the color property from object, otherwise the opening curly bracket will have the wrong color. 

Before change:
before1

After change:
after1

### DIFF
--- a/src/components/json-tree.js
+++ b/src/components/json-tree.js
@@ -25,8 +25,7 @@ export default class JsonTree extends LitElement {
         display:flex;
       }
       .json-tree {
-        background: var(--primary-color);
-        color: white;
+        background: var(--bg2);
         padding: 12px;
 
         min-height: 30px;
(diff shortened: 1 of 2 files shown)


bergander:

wparad:
Awesome I'm merging this in, it will be automatically released in 1.0.x as soon as the merge build is complete.

PR #164, opened 2023-07-14
Fix api-request leaking state when changing paths.

saraid:
Reproduction:
1. Given two path items with mutually exclusive sets of mimeTypes,
```js
{ paths: {
  "/foo": {
    "post": {
      "requestBody": {
        "content": {
          "application/octet-stream": {}
        }
      }
    },
    "put": {
      "requestBody": {
        "content": {
          "application/json": {}
        }
      }
    }
  }
}}
```
2. Navigate between them. You'll see errors in the JS console. The UI has no problems.

The error indicates that the `reqBody` is undefined, because it was searching for the incorrect `selectedRequestBodyType` in `requestBodyTypes`. This fix is just patching it so that, if it fails to find the "selected" requestBodyType, it simply picks one that at least exists.

### DIFF
--- a/src/components/api-request.js
+++ b/src/components/api-request.js
@@ -411,7 +411,8 @@ export default class ApiRequest extends LitElement {
       `;
 
     // For Loop - Main
-    const reqBody = requestBodyTypes.find(req => req.mimeType === this.selectedRequestBodyType);
+    const reqBody = requestBodyTypes.find(req => req.mimeType === this.selectedRequestBodyType) || requestBodyTypes[0];
+    this.selectedRequestBodyType = reqBody.mimeType; // Ensure that the selected type exists.
     // Generate Example
     if (this.selectedRequestBodyType.includes('json') || this.selectedRequestBodyType.includes('xml') || this.selectedRequestBodyType.includes('text')) {
       const reqBodyExamples = generateExample(


saraid:

src/components/api-request.js
```diff
@@ -411,7 +411,8 @@ export default class ApiRequest extends LitElement {
       `;
 
     // For Loop - Main
-    const reqBody = requestBodyTypes.find(req => req.mimeType === this.selectedRequestBodyType);
+    const reqBody = requestBodyTypes.find(req => req.mimeType === this.selectedRequestBodyType) || requestBodyTypes[0];
```
wparad, line 414: I'm not totally sure this is the right fix, the solution seems a bit strange. According to the problem, my guess would be we need to reset the selectedRequestBodyType on route change by watching the input parameters. Or more specifically we could watch `this.request_body` changes and the update the selected value.

wparad (changes requested):
I believe there might be a better way to handle this. As it is right now, I'm also worried about an infinite loop because we are changing the value of a property that is the one that triggered the function in the first place. I've made a suggestion inline, let me know if you aren't sure how to implement that, and I can go over an example implementation.

wparad:
Great thank you, this has been merged in #167 to the release `1.0`. If you wait a few hours a version should be released!

PR #175, opened 2023-07-30
Implement a prepare action button

### LINKED ISSUE #173: Display curl cli command before execution
Currently in the generated UI, there does not appear to be any way to view and verify the exact request that is going to be submitted (or to SEE the request to copy and run somewhere else) without clicking "Execute" which immediately invokes the request. It would be much safer and more functional if the UI would "show the invocation" and a separate click to actually invoke it.

A nice way to display this would be to just have the curl display full time (or even better a "Invocation" section, with future capability to submit additional 'formatters' similar to what Elements supports)

nneul:
Not a javascript or UI programmer, but here's at least a potential implementation that appears to work. 

Fixes #173

### DIFF
--- a/src/components/api-request.js
+++ b/src/components/api-request.js
@@ -618,8 +618,6 @@ export default class ApiRequest extends LitElement {
             </button>`
           : ''
       }
-      <button class="m-btn primary btn-prepare thin-border" part="btn btn-fill btn-try" @click="${this.onPrepareClick}">${getI18nText('operations.prepare')}</button>
-      <div class="tab-content col m-markdown">&nbsp;</div>
       <button class="m-btn primary btn-execute thin-border" part="btn btn-fill btn-try" @click="${this.onTryClick}">${getI18nText('operations.execute')}</button>
     </div>
     ${this.responseMessage === '' ? '' : this.apiResponseTabTemplate()}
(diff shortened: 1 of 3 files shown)


wparad:
I think it would make sense to take a minute to think about what the implementation could look like. Rather than a button, I would expect instead that on every change we could automatically regenerate the `curl`.

It seems like where we are going (hard to tell with the current limited changes) is that there would be a new button, but I think we can do better and avoid it by automatically generating the new curl command on every change.

I can also imagine figuring that part out could be a bit complicated, so taking a two step approach to a refactor might make the most sense. I would suggest starting with, just refactoring out the curl generation based on the `fetchOptions` from the input. If we had a fetch => curl converted, then we could call that after we call a `ui => fetch` converter.

Recommended Steps:
1. Create a UI => Fetch and Fetch => Curl converters
2. Make the `Execute` button just load the existing generated fetch command
3. Dynamically update the Fetch command based on the UI

I think that might be it, but if there is something else potentially missing we might be able to catch as we go.

What do you think about that?

nneul:
It sounds like a good design to the implementation - and one that would definitely support additional capabilties (alternative languages/displays/etc.). 

Definitely outside my area of expertise actually implementing that degree of change though.

wparad:
To worries at all. If there is anything that would still provide value however it would be refactoring the curl generation to its own function. Most of the code is already there, but it is intertwined with the execute function. If you feel up to that, it would be a great help as a first step, and then we could take it from there.

nneul:
See if that helps any.  I think it results in a bit of duplication though. It may be worth considering a function to "compile the request content into a simplified intermediate object" step and then using that from both the curl and Try and potentially others.

nneul:
However, that might also not be ideal, since using curl for example you're not using the final condensed form - you can use -F for individual form elements for example.

PR #184, opened 2023-08-19
Fix animations on schema viewer

j3rem1e:
This PR allows the "collapse" animations in schema viewer (tree/table) to be played only when the user interacts with the tree.

This fixes some glitch the first time the schema is shown or when the schema tab is clicked.

### DIFF
--- a/src/components/schema-table.js
+++ b/src/components/schema-table.js
@@ -25,6 +25,20 @@ export default class SchemaTable extends LitElement {
     if (!this.schemaHideWriteOnly || !'true false'.includes(this.schemaHideWriteOnly)) { this.schemaHideWriteOnly = 'true'; }
   }
 
+  /**
+   * @param {Map<string, object>} changedProperties Changed Properties
+   */
+  update(changedProperties) {
+    if (changedProperties.has('data')) {
+      this.interactive = false;
+    }
+    super.update(changedProperties);
+  }
+
+  updated() {
+    this.interactive = true; // Note: interactive is not a reactive property
+  }
+
   static finalizeStyles() {
     return [
       FontStyles,
@@ -76,6 +90,9 @@ export default class SchemaTable extends LitElement {
       .tr + .object-body {
         overflow: hidden;
       } 
+      .table:not(.interactive) .object-body {
+        animation-duration: 0s !important;
+      }
       .tr:not(.collapsed) + .object-body {
         animation: linear 0.2s expand-height;
       }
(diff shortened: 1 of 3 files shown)


wparad:
There is example that would showcase what is going on, so we can track the improvement here. I think I know what is up, but I would like the opportunity to make sure I'm looking at the right thing.

j3rem1e:
Yes, there is below an animation (slowed down) to show the current behavior. Every time I click on a tab, the expanded node have an animation which feel strange to me. Imo the collapse/expand animation should only be played when a user clicks on a node, not every time the view is activated.

anim_tabs

I didn't fix however the effect on the scrollbar.

src/components/schema-table.js
```diff
@@ -76,11 +90,16 @@ export default class SchemaTable extends LitElement {
...
+      .table:not(.interactive) .object-body {
+        animation-duration: 0s !important;
+      }
       .tr:not(.collapsed) + .object-body {
         animation: linear 0.2s expand-height;
       }
       .tr.collapsed + .object-body {
         animation: linear 0.2s collapse-height;
+      }
+      .tr.collapsed + .object-body {
```
wparad, line 102: This wasn't necessary, right? I assume it was an artifact of testing.

```suggestion
```

wparad (changes requested):
I don't love it, but I also don't have a better suggestion here. Let's go with this for now, but I've made two comments which would both be nice to be addressed/rectified.

src/components/schema-table.js
```diff
@@ -25,6 +25,19 @@ export default class SchemaTable extends LitElement {
     if (!this.schemaHideWriteOnly || !'true false'.includes(this.schemaHideWriteOnly)) { this.schemaHideWriteOnly = 'true'; }
   }
 
+  firstUpdated() {
+    const observer = new IntersectionObserver((entries) => {
```
wparad, line 29: It would be great if we added a comment here to explain why we are doing this (same goes for the other location). It would also be good to explain why we need a `300` timeout. What happens at 200ms, 100ms, 1ms, next tick of the event loop, that would be too soon. I think that's the part that feels the most dirty to me.

j3rem1e: Yes, I agree, it's not pretty.

The tricky part here is that it's implemented today as css animation and not transition. They play every times the dom is updated from the start of the animation definition. I'll do my best to explain it in the next commit ^^

wparad: What if we just put the class adding into the expand/collapse button clicks. That would be better if it works.

j3rem1e: I have a new working solution without setTimeout and IntersectionObserver, I will propose it to you this afternoom.

wparad (commented):
This is fantastic. Approved.

wparad (approved):

PR #185, opened 2023-08-21
Highlight matching bracket in schema-viewer

j3rem1e:
This PR add some changes on schema-viewer/json-viewer : 
- Highlight the current line on mouse-over
- Hightlight matching bracket on mouse-over
- Set the style of the description to light-fg to be more coherent to the rest of the apps and to focus more and the model
- The "writeOnly"/"readOnly" was not coherent between schema-tree and schema-table

matching_bracket

### DIFF
--- a/src/components/json-tree.js
+++ b/src/components/json-tree.js
@@ -64,12 +64,21 @@ export default class JsonTree extends LitElement {
       .tree > .open-bracket {
         margin-left: -2px;
       }
+      .open-bracket:not(.collapsed):hover > .key > .open-bracket, 
+      .open-bracket:not(.collapsed):hover + .inside-bracket-wrapper > .close-bracket,
       .open-bracket:hover {
         color:var(--primary-color);
         background-color:var(--hover-color);
         border: 1px solid var(--border-color);
       }
-
+      .open-bracket:not(.collapsed):hover + .inside-bracket-wrapper > .inside-bracket {
+        border-left: 1px solid var(--fg3);
+      }
+      .close-bracket {
+        display:inline-block;
+        border: 1px solid transparent;
+        border-radius:3px;
+      }
       .inside-bracket-wrapper {
         overflow: hidden;
       }
--- a/src/components/schema-tree.js
+++ b/src/components/schema-tree.js
@@ -76,13 +76,22 @@ export default class SchemaTree extends LitElement {
       .td.key > .open-bracket:first-child {
         margin-left: -2px;
       }
+      .object:not(.collapsed):hover > .key > .open-bracket, 
+      .object:not(.collapsed):hover + .inside-bracket-wrapper > .close-bracket {
+        color:var(--primary-color);
+      }
       .open-bracket:hover {
         color:var(--primary-color);
         background-color:var(--hover-color);
         border: 1px solid var(--border-color);
       }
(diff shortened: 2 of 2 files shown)


src/components/json-tree.js
```diff
@@ -50,12 +50,22 @@ export default class JsonTree extends LitElement {
...
       }
-
+      .open-bracket:not(.collapsed):hover ~ .inside-bracket-wrapper > .inside-bracket {
+        border-left: 1px solid var(--fg3);
+      }
+      .close-bracket {
+        display:inline-block;
+        border: 1px solid transparent;
+        border-radius:3px;
+        font-family: var(--font-mono);
```
wparad, line 67: Can we also explicitly set the font-family on the open-bracket. I know it is included in the `key`, but if feels weird without it, and who knows if we end up pulling that object into its own tag, it could break the styling at that point if we forget.

wparad (changes requested):
I'm not a fan of either of these two specific changes (the other two are good):
* Highlight the current line on mouse-over
* Set the style of the description to light-fg to be more coherent to the rest of the apps and to focus more and the model

The behavior seems messy when looking at the first one, and the second one from my perspective makes it hard to read. I know the point is to call out the text, but looking at some default styles, doesn't make this look very good. Is it just the default that is a problem or will most displays be hard to read?

This is what I see that doesn't look good:

highlight-border-broken

My suggestion is to revert those two changes, and we can try to tackle each of them in a follow up (potentially one at a time).

j3rem1e:
I have reworked this PR : Removed line hover style, the color of the text description, fixed the read/write indicator and the matching bracket color.

wparad (changes requested):
I still absolutely hate this, but I don't think it is your fault at all. The schema-tree has the open bracket nested in the `tr` which makes it impossible to only hover that. But given what it is, I think it distracting when not hovering.

So I can't approve the change to the `schema-tree` component. When the button isn't being hovered there is no way we are going to show it hovered with the left border line.

So we are going to have to remove that from the PR.

src/components/json-tree.js
```diff
@@ -64,12 +64,21 @@ export default class JsonTree extends LitElement {
...
+      .open-bracket:not(.collapsed):hover > .key > .open-bracket, 
+      .open-bracket:not(.collapsed):hover + .inside-bracket-wrapper > .close-bracket,
       .open-bracket:hover {
         color:var(--primary-color);
         background-color:var(--hover-color);
         border: 1px solid var(--border-color);
       }
-
+      .open-bracket:not(.collapsed):hover + .inside-bracket-wrapper > .inside-bracket {
+        border-left: 1px solid var(--fg3);
```
wparad, line 75: I think this one works better with the views, it feels part of the background, and I think the color works better. There is something off with fg3 for me here.
```suggestion
        border-left: 1px solid var(--bg3);
```

src/components/json-tree.js
wparad, line 1: Should we do something about this:
```
.open-bracket {
        display:inline-block;
        padding: 0 20px 0 0;
        cursor: pointer;
        border: 1px solid transparent;
        border-radius:3px;
      }
```

image


I'm almost thing we should switch `[` to `[...]` on hover, but I'm not sold on that change. It is definitely out of scope here, but if it's something you are interested in figuring out, go for it.

j3rem1e:
No problem, I close this PR.

PR #186, opened 2023-08-21
Adding highlighting to XML examples

j3rem1e:
This PR add highlighting to xml examples in response:

highlight_xml

Moreover, some css properties used by prism were not defined from the theme to the css variable. I have fixed that too.

### DIFF
--- a/src/components/api-response.js
+++ b/src/components/api-response.js
@@ -273,7 +273,7 @@ export default class ApiResponse extends LitElement {
         ? html`
           ${mimeRespDetails.examples[0].exampleSummary && mimeRespDetails.examples[0].exampleSummary.length > 80 ? html`<div style="padding: 4px 0"> ${mimeRespDetails.examples[0].exampleSummary} </div>` : ''}
           ${mimeRespDetails.examples[0].exampleDescription ? html`<div class="m-markdown-small" style="padding: 4px 0"> ${unsafeHTML(marked(mimeRespDetails.examples[0].exampleDescription || ''))} </div>` : ''}
-          <syntax-highlighter class='example-panel generic-tree border-top pad-top-8' mime-type="${mimeRespDetails.examples[0].exampleType}" .content="${mimeRespDetails.examples[0].exampleValue}" copy/>`
+          <syntax-highlighter class='example-panel generic-tree border-top pad-top-8' mime-type="${mimeRespDetails.examples[0].exampleType}" .content="${mimeRespDetails.examples[0].exampleValue}"/>`
         : html`
           <span class = 'example-panel generic-tree ${this.renderStyle === 'read' ? 'border pad-8-16' : 'border-top pad-top-8'}'>
             <select aria-label='response body example' @change='${(e) => this.onSelectExample(e)}'>
@@ -285,21 +285,13 @@ export default class ApiResponse extends LitElement {
               <div class="example" data-example = '${v.exampleId}' style = "display: ${v.exampleId === mimeRespDetails.selectedExample ? 'block' : 'none'}">
                 ${v.exampleSummary && v.exampleSummary.length > 80 ? html`<div style="padding: 4px 0"> ${v.exampleSummary} </div>` : ''}
                 ${v.exampleDescription && v.exampleDescription !== v.exampleSummary ? html`<div class="m-markdown-small" style="padding: 4px 0"> ${unsafeHTML(marked(v.exampleDescription || ''))} </div>` : ''}
-                <syntax-highlighter mime-type="${v.exampleType}" .content="${v.exampleValue}" copy/>
+                <syntax-highlighter mime-type="${v.exampleType}" .content="${v.exampleValue}"/>
               </div>  
             `)}
           </span>  
         `
       }
     `;
-  }
-
-  renderExampleTemplate(example) {
-    return html`<syntax-highlighter content-type="${example.exampleType}" content="${example.exampleValue}"/>`;
-  }
-
-  exampleValueTemplate(example) {
-    return html`<syntax-highlighter content-type="${example.exampleType}" content="${example.exampleValue}" copy/>`;
   }
 
   mimeSchemaTemplate(mimeRespDetails) {
--- a/src/components/api-response.js
+++ b/src/components/api-response.js
@@ -10,8 +10,10 @@ import InputStyles from '../styles/input-styles';
 import TabStyles from '../styles/tab-styles';
 import BorderStyles from '../styles/border-styles';
 import SchemaStyles from '../styles/schema-styles';
+import PrismStyles from '../styles/prism-styles';
 import './schema-tree';
 import './schema-table';
+import './syntax-highlighter';
 
 export default class ApiResponse extends LitElement {
   constructor() {
@@ -47,6 +49,7 @@ export default class ApiResponse extends LitElement {
       TableStyles,
       InputStyles,
       BorderStyles,
+      PrismStyles,
       css`
       .resp-head{
         vertical-align: middle;
@@ -189,7 +192,7 @@ export default class ApiResponse extends LitElement {
           </div>
           ${Object.keys(this.mimeResponsesForEachStatus[status]).length === 0
             ? ''
-            : html`  
+            : html`
               <div class="tab-panel col">
                 <div class="tab-buttons row" @click="${(e) => { if (e.target.tagName.toLowerCase() === 'button') { this.activeSchemaTab = e.target.dataset.tab; } }}" >
                   <button class="tab-btn ${this.activeSchemaTab === 'model' ? 'active' : ''}" data-tab='model'>${getI18nText('operations.model')}</button>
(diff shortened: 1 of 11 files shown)


src/components/api-response.js
```diff
@@ -308,6 +311,13 @@ export default class ApiResponse extends LitElement {
     `;
   }
 
+  exampleValueTemplate(example) {
+    if (example.exampleFormat === 'text' && example.exampleType && example.exampleType.includes('xml')) {
```
wparad, line 315:
```suggestion
    if (example.exampleFormat === 'text' && example.exampleType?.includes('xml')) {
```

src/components/api-response.js
```diff
@@ -308,6 +311,13 @@ export default class ApiResponse extends LitElement {
     `;
   }
 
+  exampleValueTemplate(example) {
```
wparad, line 314: I'm not a fan of having this inline here, we have an explicit check for `json` on the outside above the this template insertion, it's weird to then have this template explicitly check for the type again.

Also, here's something I'm not sure of, above we are checking for `exampleFormat === json`, I'm not entirely sure why it would inconsistent here. It's been a while since I looked at this, but some sort of consistency would be in how we check these.

wparad (changes requested):
This is a good change.

So, just some small things to change. If anything I think the biggest thing missing here is the answer to the question, why does the `api-request` execute response use this to display:

```html
<pre style="min-height: 60px" @copy='${() => { copyToClipboard(window.getSelection().toString()); }}'>${responseFormat
                ? html`<code>${unsafeHTML(Prism.highlight(this.responseText, Prism.languages[responseFormat], responseFormat))}</code>`
                : `${this.responseText}`
              }
              </pre>
```

does that work for XML, or does that need to be changed too? If it does work, then we should be able to do the same thing here and not explicitly pass `markup` to the Prism, but rather the responseType. And if it doesn't work then we'll need some sort of prismResolver.js to handle our fixes to prism to handle types that it doesn't support out of the box.

I don't know if I'm even making sense. thoughts?

j3rem1e:
> So, just some small things to change. If anything I think the biggest thing missing here is the answer to the question, why does the `api-request` execute response use this to display:
> 
> 
> does that work for XML, or does that need to be changed too? If it does work, then we should be able to do the same thing here and not explicitly pass `markup` to the Prism, but rather the responseType. And if it doesn't work then we'll need some sort of prismResolver.js to handle our fixes to prism to handle types that it doesn't support out of the box.
> 

Afaik it already "worked" before this PR. There is a mapping between content types and prism languages hardcoded here : https://github.com/Rhosys/openapi-explorer/blob/962e21f59eda5b535d34ba6eba2ed5bef5f01e9f/src/components/api-request.js#L564

It "worked" in the DOM, but it wasn't visible to the end user because of the css variables and styles missing. (that's why I added "language-markup" explicitly during my test, but it was already included by default in prism.js - I should remove it),

I agree it's kind of duplicated ^^
Imo, there should be some kind of "content-viewer" component which is responsable to render examples/responses + content-type, to fixes some inconsistencies. For example:
- Why is the "json-viewer" not used on response ? Moreover, the style of prism and json-viewer don't match
- Why is there no "copy button" on examples ?
- And there are conflict between the prismcss and other style, like the ".header" class - It should probably be scoped into its own component

I have hesitated to make such change, but it was just to big for this kind of PR ^^

src/components/api-response.js
j3rem1e: I agree.

The code is duplicated between the branch "only one example" and "multiple examples". But it's not exactly the same because of css classes. To minimize the change, I have only handled xml in this method, but I can try to mutualise all use cases.

wparad: I do still expect this to be a small PR and a minimized change, but I think it makes sense to move the === `json` in here, and use the same variables to decide how to format it. It's fine if it is a bunch of `if...returns`, this is probably the right time to make a subtle change to how we are rendering this, I wouldn't expect it to be large.

wparad:
> > So, just some small things to change. If anything I think the biggest thing missing here is the answer to the question, why does the `api-request` execute response use this to display:
> > does that work for XML, or does that need to be changed too? If it does work, then we should be able to do the same thing here and not explicitly pass `markup` to the Prism, but rather the responseType. And if it doesn't work then we'll need some sort of prismResolver.js to handle our fixes to prism to handle types that it doesn't support out of the box.
> 
> Afaik it already "worked" before this PR. There is a mapping between content types and prism languages hardcoded here :
> 
> https://github.com/Rhosys/openapi-explorer/blob/962e21f59eda5b535d34ba6eba2ed5bef5f01e9f/src/components/api-request.js#L564
> 
> It "worked" in the DOM, but it wasn't visible to the end user because of the css variables and styles missing. (that's why I added "language-markup" explicitly during my test, but it was already included by default in prism.js - I should remove it),
> 
> I agree it's kind of duplicated ^^ Imo, there should be some kind of "content-viewer" component which is responsable to render examples/responses + content-type, to fixes some inconsistencies. For example:
> 
> * Why is the "json-viewer" not used on response ? Moreover, the style of prism and json-viewer don't match
> * Why is there no "copy button" on examples ?
> * And there are conflict between the prismcss and other style, like the ".header" class - It should probably be scoped into its own component
> 
> I have hesitated to make such change, but it was just to big for this kind of PR ^^


I wouldn't worry about answering any of those questions yet, it's more of a question of `why` and not "we should fix it", for starters, we use the responseType in the API request - response object, I hope we can also use the responseType here to cause it to render the right thing. I think a good starting point would be to refactor the api request response to a new template and have it generate the result for the example, that feels like an easy/small change. If it isn't it would be good to know why.

> It "worked" in the DOM, but it wasn't visible to the end user because of the css variables and styles missing. (that's why I added "language-markup" explicitly during my test, but it was already included by default in prism.js - I should remove it),

In that case, if this is already mostly can be aligned then a refactored component would be even easier, and I'd be all for pulling that in. This is still small, and avoiding a digression here would help align future changes as well. It's okay if you aren't up for it, but at the very least I would hope these two rendering use almost the exact same code.

j3rem1e:
I have reworked this PR : Now all the logic behind highlighting is in only one place, included the "copy to clipboard" button

wparad:
I only took a quick look at this, but from what I've seen it looks reasonable. As soon as I get a moment, I'll do a full review.

src/components/api-response.js
```diff
@@ -308,6 +294,14 @@ export default class ApiResponse extends LitElement {
...
   }
 
+  renderExampleTemplate(example) {
+    return html`<syntax-highlighter content-type="${example.exampleType}" content="${example.exampleValue}"/>`;
+  }
+
+  exampleValueTemplate(example) {
+    return html`<syntax-highlighter content-type="${example.exampleType}" content="${example.exampleValue}" copy/>`;
+  }
+
```
wparad, line 304: These aren't used, right?

```suggestion
```

src/components/syntax-highlighter.js
```diff
@@ -0,0 +1,164 @@
...
+/**
+ * Syntax Highlighter component.
+ */
+class SyntaxHighlighter extends LitElement {
+  static get properties() {
+    return {
+      content: { type: Object },
+      language: { type: String, attribute: 'language' },
+      mimeType: { type: String, attribute: 'mime-type' },
+      copy: { type: Boolean, attribute: 'copy' }
```
wparad, line 48: We use `copy` everywhere now, as far as I can tell, so this doesn't need to be passed in.

wparad (changes requested):
Mostly just a couple of small things. So we are just about ready to get this in. The only thing left on my mind is if the `syntaxt-highlighter` really should be merged with the `json-tree`. I can't figure out a reason we would use one without the other, it's right now a "wrapper for a wrapper", but I'll let you decide if you want to merge these, I don't have a strong opinion either way.

j3rem1e:
> Mostly just a couple of small things. So we are just about ready to get this in. The only thing left on my mind is if the `syntaxt-highlighter` really should be merged with the `json-tree`. I can't figure out a reason we would use one without the other, it's right now a "wrapper for a wrapper", but I'll let you decide if you want to merge these, I don't have a strong opinion either way.

I don't know the best practices of lit-element, but to me component should be small and testable. I don't think it's better to merge them.

wparad:
> I don't know the best practices of lit-element, but to me component should be small and testable. I don't think it's better to merge them.

It's fine to have separate files that doesn't change anything, but when we have lit elements wrap each other, that creates new shadow DOM that doesn't pass through the css, so we have to reinject all our css classes which is annoying. Also things like parameters to the syntax-highlighter if it had a lot (like the api-request) then passing them all in, just to then pass them to the json-tree would also not be the best. For now, it's fine, and we'll just have to keep an eye on it, however I don't expect it to unnecessarily grow, so it shouldn't be a problem.

wparad (approved):
I'm approving this and merging it in. Nice work. As always a new version will be deployed after the build is completed.

PR #214, opened 2024-01-05
Fix display for the basic auth security scheme. fix #213

### LINKED ISSUE #213: BasicAuth Label missing
My API supports JWT token auth and basic auth, which has been configured like this in the spec:

```
    "security": [
        {
            "basicAuth": []
        },
        {
            "jwt": []
        }
    ],
```

This generally works fine in the explorer, but it seems there is some kind of label missing for the basicAuth scheme?

screenshot-localhost-2024 01 05-12_26_27
screenshot-localhost-2024 01 05-12_27_58

Not a huge deal, just looks a bit weird.

wparad:


wparad:
image

image

PR #244, opened 2024-03-08
fix? form fields not rendered

neongrau:
somehow the Generators returned were almost never producing html, except of a rare chance of <1%

this seems to render the form 100% of every time

fixes https://github.com/Authress-Engineering/openapi-explorer/issues/243


wparad:
Awesome, will use this to dive into what the generators are supposed to be doing. This is great.

wparad:
This 100% fixes the issue, merging it in.

PR #252, opened 2024-05-22
Fix setting token for security schemes other than basic and bearer

zlondrej:
The use-case I have for this is combination of session cookie and CSRF protection header.

While the cookie is handled automatically by browser, the token can be set manually in the Authentication section. However it can't be set using `setAuthenticationConfiguration` method, because the conditions only allow certain schemes to be configured.

image

Rather than preventing valid cases, the method should allow configuring any security scheme, even if it doesn't necessarily make sense.

### DIFF
--- a/src/openapi-explorer.js
+++ b/src/openapi-explorer.js
@@ -367,8 +367,6 @@ export default class OpenApiExplorer extends LitElement {
       authorizationToken = `Basic ${btoa(authorizationToken)}`;
     } else if (authorizationToken && securityObj.scheme && securityObj.scheme.toLowerCase() === 'bearer') {
       authorizationToken = `Bearer ${authorizationToken}`;
-    } else {
-      authorizationToken = null;
     }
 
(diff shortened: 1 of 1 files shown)


wparad:
Agreed, although there is no reason that a user would ever need to have a `X-CSRF-Token` to actually put in here, and even if they did this is almost certainly the wrong way to set it. There is no reason why we would assume that the Auth header value must be either `Basic` or `Bearer`. So it makes sense to pull in this change.

wparad:
Published in version `2.2.710`

zlondrej:
Well, that was way faster than I expected, I didn't even manage to write the changelog. :grimacing:

wparad:
> Well, that was way faster than I expected, I didn't even manage to write the changelog. 😬

Honestly I thought about it, but I don't think anyone has been waiting for this feature. Most people don't expose the token input, so it isn't he highest used feature OR they are using Bearer/Basic input. So I didn't think we needed to take that extra step here.

PR #253, opened 2024-05-24
Schema representation improvements

zlondrej:
I made some changes to the way schemas are visualized to better fit my use cases, but I think they're more general use cases and are in-line with what OpenAPI is describing.

image

I also added new `schema-compact-single-x-of-option` option for more compact display of single item `xxxOf`. It would be better to move the schema link to the type column, but internal representation is a too messy for that and I can't spend that much extra time on it, as that would require substantial rewrite.

image

Additional details can be found in the documentation and changelog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,15 @@ This package follows standard semvar, `<major>.<minor>.<build>`. No breaking cha
 * Prevent making requests when required path parameters are not specified.
 * Automatically retry fetching the spec if it doesn't work for any reason.
 * Fix array description display preference to be first--The Array description, and then only second--The array item schema description, so that the "more specific" description wins, even though the "items" description is more deeply nested. This aligns to the expected behavior of preference in the json schema.
+* Treat `allOf` the same way as `oneOf` and `anyOf` instead of trying to merge them as a single object,
+  despite the not even necessarily being objects.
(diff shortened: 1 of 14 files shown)


wparad:
I'm concerned that there is just too much going on in this PR in order to test, validate, and review it effectively. Can you split this and start with just the smallest one change you consider to be most important, and then we can work on it from there?

Right now I have too many questions and the individuals discussions are going to get conflated as we try to discuss each one.

wparad (changes requested):
Re comment, please split.

zlondrej:
Well, I was expecting this a bit but it was worth a try. I will try to split them once I have some more time to spare on this.

wparad:
> Well, I was expecting this a bit but it was worth a try. I will try to split them once I have some more time to spare on this.

Sorry. Fwiw, there about 70% which is great, and 30% I have concerns with. Maybe PR everything that's refactoring, and open issues for the functional changes where we can discuss each one. In general, we like to avoid:
* input parameters - they add confusion
* Open transparency - think "as a user of the spec" rather than "as the developer of the spec"
* Undetermined form - a feature can be great, but if we can't figure out how to display it perfectly then, it will be a struggle for users to not have unnecessary cognito overhead

Instead optimize for:
* sane default
* usability over features
* clarity over transparency

wparad:
Closing in favor of smaller PRs. Looking forward to them.

PR #272, opened 2025-05-11
Persist the api key in LocalStorage

olivergg:
This is just a draft to explore how to persist the api token in  LocalStorage, the code is not ready to be merged

### DIFF
--- a/src/templates/security-scheme-template.js
+++ b/src/templates/security-scheme-template.js
@@ -30,6 +30,7 @@ function onApiKeyChange(apiKeyId, e) {
   }
 
   securityObj.finalKeyValue = apiKeyValue;
+  localStorage.setItem('openapi-explorer', JSON.stringify({ finalKeyValue: apiKeyValue }));
   this.requestUpdate();
 }
 
@@ -40,6 +41,7 @@ function onClearAllApiKeys() {
     v.value = '';
     v.finalKeyValue = '';
   });
+  localStorage.removeItem('openapi-explorer');
   this.requestUpdate();
 }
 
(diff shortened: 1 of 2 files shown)


wparad:
It can be a good idea to store the credential information for a user if this makes for them on your site.

And the OpenAPI explorer already offers ways to programmatically inject in keys from your UI without needing to have this functionality embedded in the explorer code.

Attempting to have the explorer store the keys itself is not a good idea as it means taking more responsibility in how to interact with this saved data, there are too many ways for this to go wrong:
* User has rejected the user of local storage
* LocalStorage implementation is broken
* User wants to overwrite the localstorage version
* Implementor of the library wants to store the value in localStorage in a different place
* Implementor doesn't want the value stored.

And the list goes on, as such I am closing this PR. If you need assisstance using the methods provided by the library to automatically inject in the values. [Check out the documentation on how to do this](https://github.com/Authress-Engineering/openapi-explorer/blob/release/2.2/docs/documentation.md#library-api)

The relevant method is:
```js
setAuthenticationConfiguration(securitySchemeId, { token, clientId, redirectUri })
```

And you can wrap that call with reading from your localstorage.

If for some reason this is insufficient for your needs, can you open an issue ticket so that it can be discussed what the problem is that needs a resolution? (Jumping straight into a PR skips this necessary step)

olivergg:
Thank you @wparad, this approach is even better as it allows me to attach an event to a specific endpoint to retrieve the token and set it programmatically (in addition to storing it in local storage for convenience when the user reloads).

Would you mind if I opened another PR to improve the documentation for this use case? Because I find the documentation to be a bit scarce on how to use setAuthenticationConfiguration. It could be useful for people coming from rapidoc used to the "persistAuth" option.

wparad:
Oh definitely, please feel free to improve the documentation, that would definitely help show us where the challenge points might be in implementing it and if we need to better extend the functionality to hit every edge case.

PR #275, opened 2025-05-11
Fix keyboard navigation and form input label accessibility issues

### LINKED ISSUE #274: openapi-explorer fails standard accessibility checks
I was looking at this as a replacement for swagger UI, which has its own accessibility failures. However, this utility is also extremely inaccessible in some very basic (and hopefully easily fixable) ways.

- Rather than using traditional links or buttons, every menu item is a div, with no aria roles, and no tabindexes, and thus no ability to navigate the menus with keyboard controls
- Focus styles for most buttons are nonexistant, so the user cannot tell what is in focus when they _are_ able to use the keyboard to navigate
- Pretty much every text input is missing a label

There are probably other, less obvious accessibility issues, but these are the major ones I can see immediately. Using accessibility tools and checkers on the code created by this tool, and trying to navigate the built screen using keyboards, would go a long way toward discovering these and any additional problems.

Currently, anyone using this tool on a website meant for public access in countries with legal web accessibility requirements (such as the United States) runs the risk of being sued.

ahamelers:
Closes https://github.com/Authress-Engineering/openapi-explorer/issues/274


src/components/api-request.js
```diff
@@ -158,7 +158,7 @@ export default class ApiRequest extends LitElement {
         return html`
           <tr> 
             <td colspan="1" style="width:160px; min-width:50px; vertical-align: top">
-              <div class="param-name ${generatedParamSchema.deprecated ? 'deprecated' : ''}" style="margin-top: 1rem;">
+              <div class="param-name ${generatedParamSchema.deprecated ? 'deprecated' : ''}" style="margin-top: 1rem;" id="request-${paramName}">
```
wparad: Assuming this is only for the aria, can we rename this to be more explanatory, maybe:
```suggestion
              <div class="param-name ${generatedParamSchema.deprecated ? 'deprecated' : ''}" style="margin-top: 1rem;" id="request-${paramName}-label">
```

src/components/api-request.js
```diff
@@ -493,6 +496,7 @@ export default class ApiRequest extends LitElement {
                   @input="${() => { this.computeCurlSyntax(); }}"
                   class = "textarea request-body-param-user-input"
                   part = "textarea textarea-param"
+                  aria-label = "Request body"
```
wparad: I think this needs to use the I18n framework to resolve the value, there might already be a mapping for this already, but I don't recall.

src/styles/input-styles.js
```diff
@@ -111,6 +118,14 @@ input[type="password"]:active {
...
 }
 
+a:focus-visible,
+section .nav-bar-path:focus-visible span {
+  outline:thin solid var(--secondary-color);
+}
+section .nav-bar-path:focus-visible span {
+  outline-offset: 2px;
+}
+
```
wparad: Can you share what this is going for? Also I believe this might be necessary to be in the `nav-styles.js` and not in here.

src/templates/navbar-template.js
```diff
@@ -39,7 +39,7 @@ export function expandCollapseAllComponents() {
 /* eslint-disable indent */
 export default function navbarTemplate() {
   return html`
-  <nav class='nav-bar ${this.renderStyle}' part="section-navbar">
+  <nav class='nav-bar ${this.renderStyle}' part="section-navbar" aria-label="API navigation">
```
wparad: I'm assuming this needs the i18n resolution as well.

src/templates/navbar-template.js
```diff
@@ -62,18 +62,19 @@ export default function navbarTemplate() {
     ${html`<nav class='nav-scroll' part="navbar-scroll">
       ${(this.hideInfo || !this.resolvedSpec.info)
         ? ''
-        : html`<div class='nav-bar-info' id='link-overview' data-content-id='overview' @click = '${(e) => this.scrollToEventTarget(e, false)}'>
+        : html`<div class='nav-bar-info' id='link-overview' data-content-id='overview' @click = '${(e) => this.scrollToEventTarget(e, false)}' role="link" tabindex="0" @keydown = '${(e) => { if (['Enter', 'Space'].includes(e.key)) { e.target.click(); }}}';
```
wparad: Question for my own understanding, `Space` seems wrong to me, Space usually controls the scrollbar, is there a reason we adding `Space` to each of these, `Enter` makes sense to me.

src/styles/input-styles.js
```diff
@@ -24,6 +24,11 @@ export default css`
   background-color: var(--primary-color);
   color: var(--primary-btn-text-color);
 }
+.m-btn.primary:hover,
+.m-btn.primary:focus-visible {
+  color: var(--primary-color);
+  background-color: var(--primary-btn-text-color);
+}
```
wparad: I think this actually breaks the experience:
Peek 2025-05-12 16-08


I don't know what the correct fix is, it looks like there might be a mistake here in the underlying library not marking these as `active`.


My suggestion is going to be revert this for the PR, and then we can come up with a separate solution out of band so the rest of these changes can be merged in, unless you have an obvious solution that will fix this.

One idea that comes to mind is changing the border/text color like we do with the nav items, without changing the background. Thoughts?

wparad (changes requested):
Mostly good improvements. I do have a couple of a questions, and one thing that will need changing before merge (details in relevant comment).

src/styles/input-styles.js
ahamelers: The hover state for the non-active buttons makes them look active (currently selected). The hover state for the selected buttons, currently, does not exist. The obvious quick thing to me was to reverse it in the same way the selected one was reversed, but if I were designing my own experience from scratch each button style (selected and normal) would have its own hover style that *did not overlap* with the active/selected style.

src/styles/input-styles.js
ahamelers: The goal of this is to add a focus style, in the style of links, to nav-bar links that appear on tag pages. Outlines are the default for focused elements, including links (to make links focusable and see this in action, you may need to adjust your browser settings and turn on the ability to navigate via keyboard). Originally I was applying this style to buttons as well, then realized outlines seemed to have been turned off of buttons purposefully, but that's how it ended up in this file. Moving it makes sense!

src/templates/navbar-template.js
ahamelers: When tabbing through focusable elements using keyboard navigation, the space bar activates buttons (and some other inputs). It was difficult to choose whether these elements should have the link or button role—they're an odd combination, since clicking them doesn't _really_ navigate to a new page, but does an action (like a button), but the action does ultimately adjust the window URL to an address for uniquely linkable content (like a link). Although I decided to use the link role for these, I thought it was safest to allow both link and button activation methods via the keyboard.

wparad: In the interest of potentially not breaking the usability here in the future, I like the link, that resonates. Can we remove the `Space` for now, and if there is push back on desired usage here we can add it in later without any breaking changes.

src/components/api-request.js
```diff
@@ -175,6 +175,7 @@ export default class ApiRequest extends LitElement {
                   <tag-input class="request-param" 
                     autocomplete="on"
                     id = "request-param-${paramName}"
+                    aria-labelledby = "request-${paramName}"
```
wparad: We update the `id` so this should be:

```suggestion
                    aria-labelledby = "request-${paramName}-label"
```

Same goes for all the other ones.

src/styles/input-styles.js
```diff
@@ -43,9 +38,9 @@ export default css`
 .m-btn.tiny { padding:5px 6px; }
 .m-btn.circle { border-radius: 50%; }
 .m-btn:hover,
-.m-btn:focus-visible { 
-  background-color: var(--primary-color);
-  color: var(--primary-btn-text-color);
+.m-btn:focus-visible {
+  border-color: var(--secondary-color);
+  color: var(--secondary-color);
```
wparad: I don't really like the text color changing but the border changing seems appropriate, I'm going to make this change so we can get this in.

src/styles/tab-styles.js
```diff
@@ -38,6 +38,15 @@ export default css`
 .tab-btn:hover {
   color:var(--primary-color);
 }
+.tab-btn.active:hover {
+  color:var(--secondary-color);
+}
```
wparad: I'm removing this, because I don't understand what the purpose it. Right now it looks like it is wrongly conveying that the active active can be clicked, it cannot be as far as I am aware.

src/templates/navbar-template.js
wparad: Updated.

src/components/api-request.js
wparad: Updated.

wparad (approved):
I've made some small tweaks here based on my feedback and some slight style improvements. Feel free to take one last look and make sure that I didn't break anything that needed to be changed.

If nothing stands out, I'll merge it later today.

wparad (approved):

wparad:
Latest updates are fine, next time while entirely appreciated, let's try to avoid unnecessary scope creep (ie additional translations)

PR #277, opened 2025-05-12
Fix switching between request body examples

### LINKED ISSUE #276: Cannot switch between request body examples
If there are multiple request body examples, a drop down appears allowing you to select between them. But choosing a different example from the drop down **does not** change the example shown in the textarea. 

Here's some example YAML that doesn't work.

```
paths:
  /:
    post:
      summary: Test post
      requestBody:
        required: true
        content:
          application/json:
            schema:
              type: object
              properties: 
                text:
                  type: string
            examples:
              test1:
                value: {text: 'test1'}
              test2:
                value: {text: 'test2'}
```

See the screenshots. You can see after selecting 'Test 2', the example shown does not change.

Image
Image

This is unlike response body examples, which work as expected.

ahamelers:
Closes https://github.com/Authress-Engineering/openapi-explorer/issues/276

### DIFF
--- a/src/components/api-request.js
+++ b/src/components/api-request.js
@@ -384,7 +384,7 @@ export default class ApiRequest extends LitElement {
     this.selectedRequestBodyExample = e.target.value;
     const exampleDropdownEl = e.target;
     window.setTimeout((selectEl) => {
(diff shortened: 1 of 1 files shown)


wparad:
Amazing, thank you.

PR #284, opened 2025-07-22
Allow description in schema elements in additon to the description on the level above.

Xarno:
The kotlin openapi generator put the description for query parameter inside the schema element. So this change allows to render the text from there.

Example:
The change adds the text marked by blue arrows and maintains the purple one.
Render more texts

### DIFF
--- a/src/components/api-request.js
+++ b/src/components/api-request.js
@@ -155,6 +155,14 @@ export default class ApiRequest extends LitElement {
 
       const rowGenerator = ({ name: paramName, description: paramDescription, required: paramRequired }, generatedParamSchema) => {
         const displayAllowedValuesHints = (generatedParamSchema.type === 'object' || generatedParamSchema.type === 'array') && generatedParamSchema.allowedValues;
+        let description = paramDescription ?? '';
+        if (generatedParamSchema.description?.length > 0 && generatedParamSchema.description !== paramDescription) {
+          if (description.length > 0) {
+            description += '<br>';
+          }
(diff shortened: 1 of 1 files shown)


wparad:
Before we make any change, we'll need to validate that this indeed valid syntax according to the OAS 3.1+? Can you link me to where `schema.description` is defined and where its semantics are specified, so that we can be sure this is the appropriate implementation.

Xarno:
Thanks for the pointer to look again and sorry to have bothered you. 
It turn's out that not the `@Schema` annotation of the generator needs to be used at a parameter — even if it have a description field — but the `@Parameter` annotation. Then the text moves from the „blue“ position to the „purple“ one and is rendered.

That we have sometimes two texts is our problem, and I have a way to join them if needed.

wparad:
For what it's worth (as well as anyone else that sees this in the future). You can dynamically make adjustments to your spec using this library, that go into affect **before render**  by doing something like this using the `onSpecLoaded` hook for the library:

```js
recursiveFindAndUpdateAllDescriptionFields(param) {
  if (!param || typeof param !== 'object') {
    return;
  }

  if (Array.isArray(param)) {
    param.map(p => recursiveFindAndUpdateAllDescriptionFields(p));
    return;
  }

  if (param.description && param.schema.description) {
    param.descrption += `<br />${param.schema.description}`;
    return;
  }
}

onSpecLoaded(data) {
  recursiveFindAndUpdateAllDescriptionFields(data.detail);
}
```

PR #292, opened 2025-12-17
navigate to complete url instead of only fragment/hash

toy:
Using only fragment works completely fine unless [base tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/base) is added in head (which makes also pushState urls be relative to it instead of to the current url). Fix by constructing complete url relative to current location.

### DIFF
--- a/src/utils/common-utils.js
+++ b/src/utils/common-utils.js
@@ -149,7 +149,10 @@ export function replaceState(rawElementId) {
   const query = new URLSearchParams(currentQuery);
   query.delete('route');
   const newQuery = query.toString().length > 1 ? `${query.toString()}&route=${elementId}` : `route=${elementId}`;
-  window.history.pushState(null, null, `#${currentNavigationHashPart}?${newQuery}`);
+
+  const fragment = `#${currentNavigationHashPart}?${newQuery}`;
+  const url = new URL(fragment, window.location.href);
+  window.history.pushState(null, null, url.href);
 }
 
 export function toMarkdown(markdownStringRaw) {


wparad:
Can you provide a working reproduction of this problem?

toy:
Following would be a minimal reproduction, the fragment will be prefixed with `/not/the/path/you/expected/here`:
```html
<!doctype html>
<html>
  <head>
    <meta charset="utf-8">
    <title>OpenAPI Explorer Development Test</title>
    <script src="https://unpkg.com/openapi-explorer/dist/browser/openapi-explorer.min.js" type="module" defer></script>
    <base href="/not/the/path/you/expected/here" />
  </head>
  <body>
    <openapi-explorer spec-url="https://api.authress.io/v1"></openapi-explorer>
  </body>
</html>
```

wparad:
to complete the reproduction can please include these necessary additional pieces:
* How do you trigger the bug?
* What do you expect to happen?
* What actually does happen?

As soon as we get those, we can verify this fix solves the problem. Thanks!

toy:
> * How do you trigger the bug?

Having the openapi-explorer rendered at a path on a page with base tag in head with href attribute (by default set to `/` which is not equal to the path of openapi-explorer page). Example at https://community.openproject.org/api/docs

> * What do you expect to happen?

Reloading the page stays on the openapi explorer page and at the same location inside

> * What actually does happen?

The path from href of base tag is loaded with fragment set by openapi-explorer

wparad:
Cool, we support this change.

PR #296, opened 2026-03-12
Add configurable load-timeout to show error instead of endless spinner

dennisoelkers:
When loading a spec from a slow/unreachable URL, ProcessSpec can hang forever because openapi-resolver's abort handler swallows the fetch rejection. Wrap ProcessSpec in Promise.race with a configurable timeout so the existing error display path is triggered.

This PR adds a `load-timeout` attribute (default: 5000ms) that controls how long to wait before showing a timeout error. Usage:

```html
  <openapi-explorer spec-url="..." load-timeout="10000"></openapi-explorer>
```

### DIFF
--- a/src/openapi-explorer.js
+++ b/src/openapi-explorer.js
@@ -155,7 +155,8 @@ export default class OpenApiExplorer extends LitElement {
       // Internal Properties
       loading: { type: Boolean }, // indicates spec is being loaded
       showAdvancedSearchDialog: { type: Boolean },
-      advancedSearchMatches: { type: Object }
+      advancedSearchMatches: { type: Object },
+      loadTimeout: { type: Number, attribute: 'load-timeout' }
     };
   }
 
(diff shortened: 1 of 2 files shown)


wparad:
If the resolver in swallowing something that it shouldn't and that prevents us from knowing about the issue, the correct fix is to update the resolver, and not wrap a hack here. Please file an issue in that repo instead.

dennisoelkers:
@wparad: Valid point. I think it still makes sense (and would help us) to make the initial load timeout configurable. Right now it is hardcoded to 5s, which is not enough in some cases for us (initial load takes longer). So can I reopen this PR in a modified version just for the configurable load timeout?

wparad:
I'm not seeing where a timeout being set in this repository, do you?

dennisoelkers:
Never said that, but if it is supposed to be configurable, we would need to declare the parameter in this repository, wouldn't we?

On deeper inspection, it turned out that the swallowing of the timeout error seems to be caused by `@apidevtools/json-schema-ref-parser` which is used in a rather old version (`v9.0.9`, it's from 2021) and the most recent one is fixing this issue + has a higher timeout of 60s by default. Would it be an option to update the use of it in `openapi-resolver.js`? A quick look at the APIs looks like they are pretty compatible.

wparad:
If they are compatible, then absolutely. Feel free to make that PR, and someone can review it.

PR #306, opened 2026-07-14
Keyboard navigation, focusing, and disclosure elements

ahamelers:
See https://github.com/Authress-Engineering/openapi-explorer/issues/305

According to accessibility auditors:

- All clickable elements must also be navigable by keyboard. 
- When clicking an element changes the main page content, focus should move to that content.
- Focus should never be lost.
- Elements that are hidden visibly and contain focusable elements should also be hidden to screen readers, so the focus does not disappear inside them.
- When clicking something hides or shows sections of content, aria-expanded should accurately reflect this.

I've handled some major elements (adding a treegrid and many tablists) in a couple separate pull requests. This one covers general keyboard navigation and accurate use of aria-controls and aria-expanded.

### DIFF
--- a/src/components/json-tree.js
+++ b/src/components/json-tree.js
@@ -75,10 +75,12 @@ export default class JsonTree extends LitElement {
         animation-duration: 0s !important;
       }
       .open-bracket:not(.collapsed) + .inside-bracket-wrapper {
+        visibility: visible;
         animation: linear 0.2s expand-height;
       }
       .open-bracket.collapsed + .inside-bracket-wrapper {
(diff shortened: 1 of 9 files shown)


wparad:
There are too many changes to review here. Please split the changes into individual PRS and one PR should be submitted at a time. Multiple PRs will be rejected. In each PR include what exactly the problem is, why the current implementation is insufficient and what must change t support it your needs.

I'm closing this PR as per the above and posting this same message on all other PRs.

Please focus on one PR until that PR is approved and merge. We cannot support reviewing multiple PRs at the same time.

PR #309, opened 2026-07-14
Use tab roles for tabs

ahamelers:
See https://github.com/Authress-Engineering/openapi-explorer/issues/305

Tab panels and content should use the recommended aria roles and keyboard navigation.


wparad:
There are too many changes to review here. Please split the changes into individual PRS and one PR should be submitted at a time. Multiple PRs will be rejected. In each PR include what exactly the problem is, why the current implementation is insufficient and what must change t support it your needs.

I'm closing this PR as per the above and posting this same message on all other PRs.

Please focus on one PR until that PR is approved and merge. We cannot support reviewing multiple PRs at the same time.

PR #313, opened 2026-07-15
Accessibility fixes for copy button

ahamelers:
Currently, there are several accessibility issues with the copy button:

1. The function works in such a way that the focus is removed from the button, stranding the keyboard/screen reading user
2. If there are multiple copy buttons on the same page, they all have the same text and so cannot be discerned from one another by the screen reader user
3. The 'copied' success status update is not announced to the screen reader user.

This PR fixes these issues by:

1.  Updating the copy function to use the clipboard API and no longer change focus
2. Adding an aria-label to each copy button to make it clear to the screen reader user what they are copying
3. Updating an aria status element so the 'copied' success status update is announced

Also note that [scrollable regions on a page must be able to receive tab focus](https://www.w3.org/WAI/standards-guidelines/act/rules/0ssw9k/), and stuff that's focussable should be labeled and given a role if possible, so this PR also adds a label, role, and tabindex to the scrollable region that the copy button copies.

### DIFF
--- a/src/components/api-request.js
+++ b/src/components/api-request.js
@@ -659,14 +659,14 @@ export default class ApiRequest extends LitElement {
             </div>`
           : html`
             <div class="tab-content col m-markdown" style="flex:1; display:${this.activeResponseTab === 'response' ? 'flex' : 'none'};" >
-              <syntax-highlighter style="min-height: 60px" mime-type="${this.responseContentType}" .content="${this.responseText}" aria-label="Response text"/>
+              <syntax-highlighter style="min-height: 60px" mime-type="${this.responseContentType}" .content="${this.responseText}" .label="Response text"/>
             </div>`
         }
         <div class="tab-content col m-markdown" style="flex:1;display:${this.activeResponseTab === 'headers' ? 'flex' : 'none'};" >
-          <syntax-highlighter style="min-height: 60px" language="http" .content="${this.responseHeaders}" aria-label="Response headers"/>
+          <syntax-highlighter style="min-height: 60px" language="http" .content="${this.responseHeaders}" .label="Response headers"/>
         </div>
         <div class="tab-content m-markdown col" style="flex:1;display:${this.activeResponseTab === 'curl' ? 'flex' : 'none'};">
-          <syntax-highlighter style="min-height: 60px" language="shell" .content="${curlSyntax.trim()}" aria-label="Request example"/>
+          <syntax-highlighter style="min-height: 60px" language="shell" .content="${curlSyntax.trim()}" .label="Request example"/>
         </div>
       </div>`;
   }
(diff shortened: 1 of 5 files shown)


src/components/syntax-highlighter.js
```diff
@@ -96,7 +97,7 @@ class SyntaxHighlighter extends LitElement {
   }
 
   render() {
-    return this.renderCopyWrapper(this.renderHighlight());
+    return this.renderCopyWrapper(this.renderHighlight(), this.label?.toLowerCase());
```
wparad: We require the label everywhere for this to work, we shouldn't need the `?.` right?

src/utils/common-utils.js
```diff
@@ -32,26 +33,20 @@ export function copyToClipboard(copyData, eventTarget) {
...
-  textArea.focus();
-  textArea.select();
   try {
-    document.execCommand('copy');
-    const btnEl = eventTarget?.target;
+    await window.navigator.clipboard.writeText(data);
     if (btnEl) {
+      const label = btnEl.getAttribute('aria-label');
       btnEl.innerText = getI18nText('operations.copied');
+      btnEl.setAttribute('aria-label', getI18nText('operations.copied'));
```
wparad: Is this really the correct thing to do for a screen reader? Setting the aria-label and changing it? That feels wrong to me, but I don't really understand how aria-labels work. What's your thinking here?

ahamelers: It's a bit complicated because the button itself contains the status update text for the success ('copied'). We want to let users know the 'copied' success status, so there's an aria-live region on the button. But screen readers will never read the actual content of the button if aria-label is also set, they will always read that instead. So we have to update the aria-label as well for the status to be announced.

wparad: I refuse to believe this is right way to solve this problem for screen readers. Why would a screen reader announce the button label change?

Multiple llms suggested to me to have a separate hidden div with the text that can be targeted for update instead of this here. Because actually I worry that since we are updating the button text twice which means that it could actually display both the "copied" message and then again the "label" message which not correct.

Is that accurate, IDK, but that feels better than this implementation.

src/components/api-request.js
```diff
@@ -659,14 +659,14 @@ export default class ApiRequest extends LitElement {
             </div>`
           : html`
             <div class="tab-content col m-markdown" style="flex:1; display:${this.activeResponseTab === 'response' ? 'flex' : 'none'};" >
-              <syntax-highlighter style="min-height: 60px" mime-type="${this.responseContentType}" .content="${this.responseText}"/>
+              <syntax-highlighter style="min-height: 60px" mime-type="${this.responseContentType}" .content="${this.responseText}" .label="Response text"/>
```
wparad: Almost forgot, all of these labels need to be i18n keys instead of hard coded text because we support multiple languages.

wparad (approved):

